PR #2, opened 2020-06-06
Added a snippets section, with some snippets

torch2424:
Discussed this offline with @dcodeIO :smile: 

Wanted to have some really super small, copy-pastable pattern / AS syntax examples in the docs :smile: Especially something searchable (as I often find myself looking for an example in the docs, but can only find it's limitations or more reference documentation). 

Thus, I created this little easy-to-add snippet section, where we can write really self explanatory examples, that just show off how a small thing could be done in AS. I didn't want the PR to be super huge. So I only added a few for now, but my plan is to also add examples relating to these tests/topics:

* https://github.com/AssemblyScript/assemblyscript/blob/master/tests/compiler/scoped.ts
* https://github.com/AssemblyScript/assemblyscript/blob/master/tests/compiler/instanceof.ts
* https://github.com/AssemblyScript/assemblyscript/blob/master/tests/compiler/inlining.ts
* https://github.com/AssemblyScript/assemblyscript/blob/master/tests/compiler/getter-call.ts
* https://github.com/AssemblyScript/assemblyscript/blob/master/tests/compiler/class-extends.ts
* https://github.com/AssemblyScript/assemblyscript/blob/master/tests/compiler/class.ts
* https://github.com/AssemblyScript/assemblyscript/blob/master/tests/compiler/call-super.ts
* https://github.com/AssemblyScript/assemblyscript/blob/master/tests/compiler/call-optional.ts
* Object Literals
* Function Types / Passing Functions
* Pointers (alloc, free, retain, release)
* Triple Equals vs. Double Equals

# Screen shots

Screenshot from 2020-06-05 17-51-00
Screenshot from 2020-06-05 17-49-57
Screenshot from 2020-06-05 17-48-39
Screenshot from 2020-06-05 17-48-28
Screenshot from 2020-06-05 17-48-10


src/snippets/ternary.md
```diff
@@ -0,0 +1,35 @@
...
+<script>
+const jsLog = console.log;
+function domConsoleLog() {
+  let args = [...arguments];
+  jsLog.apply(this, args);
+  args.forEach(arg => {
+    document.body.innerHTML += `<div>Log: ${args}</div>`;
+  }); 
+}
+console.log = domConsoleLog;
```
MaxGraey:
```suggestion
const jsLog = console.log;
function domConsoleLog(...args) {
  jsLog(...args);
  args.forEach(arg => {
    document.body.innerHTML += `<div>Log: ${args}</div>`;
  }); 
}
console.log = domConsoleLog;
```
wdyt?

src/snippets/ternary.md
```diff
@@ -0,0 +1,35 @@
+---
+description: Ternary Snippet
+---
+
+# Ternary
+
+Using Ternary in AssemblyScript
+
+```editor
+#!optimize=size&runtime=full
```
dcodeIO: Since the HTML does not utilize runtime helpers, we can have a much smaller wat file here by changing

```suggestion
#!runtime=half
```

Also, optimize defaults to `speed`, which yields the same result for small snippets and compiles a little faster due to skipping `--converge`.

src/snippets/switch.md
```diff
@@ -0,0 +1,57 @@
+---
+description: Switch Statement Snippet
+---
+
+# Switch
+
+Using `switch` statements in AssemblyScript
+
+```editor
+#!optimize=size&runtime=full
```
dcodeIO:
```suggestion
#!runtime=half
```

src/snippets/switch.md
```diff
@@ -0,0 +1,57 @@
...
+export function switchSurprise(a: i32): i32 {
+  let response = -1;
+  switch(a) {
```
dcodeIO:
```suggestion
  switch (a) {
```

Also, typical code style on the docs is omitting semicolons since it is a bit easier to read and often considered "more modern". Differs from the compiler, where the greater goal is to be better safe than sorry :)

src/snippets/switch.md
```diff
@@ -0,0 +1,57 @@
...
+    case 1:
+      response = 100;
+      break;
+    // Cases can also use braces
+    case 2: {
+      response = 200;
+      break;
+    }
+    case 3:
```
dcodeIO: This appears to mix braced / non-braced cases. Is this intended as part of the example? If not, I suggest to use braces everywhere so an eventual `let` or `const` variable declaration becomes correctly bound to the case, not the switch.

src/snippets/switch.md
```diff
@@ -0,0 +1,57 @@
+---
+description: Switch Statement Snippet
+---
+
+# Switch
+
+Using `switch` statements in AssemblyScript
```
dcodeIO: Could mention that switch conditions currently implicitly convert to `u32`, i.e. switching over  `string`s or similar is not yet supported. If using `i32`s, a more technical detail would be that using values greater than or equal to zero has better codegen, because `br_table`s are unsigned, so a `-1` for example would not tablify well with otherwise positive values (it does work, though, but code is not optimal).

src/snippets/ternary.md
MaxGraey: Or something like this:
```js
const log = console.log;
console.log = (...args) => {
  log(...args);
  let str = '';
  args.forEach(arg => {
    if (typeof arg == 'object') {
      str += `${JSON.stringify(arg, null, 2)}<br/>`;
    } else {
      str += `${arg}<br/>`;
    }
  }
  document.body.innerHTML += `<div>Log: ${str}</div>`;
}
```

src/snippets/class.md
```diff
@@ -0,0 +1,58 @@
+---
+description: Class / Classes Snippet
+---
+
+# Class
+
+A snippet on using Classes in AssemblyScript, and assigning instance vs. static properties.
+
+```editor
+#!optimize=size&runtime=full
```
dcodeIO:
```suggestion
#!runtime=full
```

As in the earlier comment, this compiles a little faster due to skipping `--converge`. I guess using the full runtime here is reasonable, as we are dealing with managed objects here under the hood, even though the HTML does not receive these. Perhaps retaining/releasing a managed object, preferably a string so it doesn't become too complex, would be a good last case in the HTML?

src/snippets/extending-classes.md
```diff
@@ -0,0 +1,68 @@
...
+description: Extending classes, object orientated programming snippet
+---
+
+# Extending Classes
+
+Extending classes, and general Object Orientated Programming (OOP) in AssemblyScript
+
+
+```editor
+#!optimize=size&runtime=full
```
dcodeIO:
```suggestion
#!runtime=full
```

Whether the full runtime is good here or not again depends a bit. With it, the wat file is considerably larger, but as we are dealing with classes here as well, one gets the exported runtime helpers which might be useful. Perhaps a `string` case in the HTML?

src/snippets/handling-null.md
```diff
@@ -0,0 +1,66 @@
...
+  if (isAble) {
+    return myValue;
+  } else {
+    return null;
+  }
+}
+
+export function positiveAddWithMyValue(a: i32): i32 {
+  let myValue = getMyValue(a > 0);
+  if (myValue == null) {
```
dcodeIO: This is more of a comment than an actual suggestion, since writing it this way is typical, even though it behaves a little different from `===`. It does not in this case, because `MyValue` does not declare an `==` overload, but if it would, `===` (identity equals) can be precomputed while `==` (calls overload function) cannot. Only relevant if eliminating branches with constant values is the goal, and a class with such an overload is the subject (i.e. a `string`). Thought I mention in case you have an idea how to convey this in a way that it doesn't become "too much" for a snippet.

src/snippets/index.md
```diff
@@ -0,0 +1,12 @@
+# Snippets
```
dcodeIO: Wondering if the snippets could be a single page, instead of multiple, given that these are pretty much just an editor? Unless you plan to add more content above or below the editor for each snippet that is. The sidebar would appear as the same if each snippet follows a `##` heading, and the search would pick these up still. The two scenarios on my mind are:

* Each snippet has a bit of prose to it above or below the editor, explaining what it does or giving context, then multiple pages would be fine as pages wouldn't look too empty.
* It's just a headline and the editor, where a single page would feel like "richer" documentation (for the lack of a better word)

torch2424:
Thanks for the suggestions @MaxGraey and @dcodeIO ! :smile: 

I looked over all the suggestions, and I will do all of them once I get the chance :+1: 

But in particular:

> Wondering if the snippets could be a single page, instead of multiple, given that these are pretty much just an editor? Unless you plan to add more content above or below the editor for each snippet that is. The sidebar would appear as the same if each snippet follows a ## heading, and the search would pick these up still. 

Totally! I don't mind having one super long page of snippets, which is how [30 seconds of CSS used to do it](https://github.com/atomiks/30-seconds-of-css)! And to your point of:

> Each snippet has a bit of prose to it above or below the editor, explaining what it does or giving context, then multiple pages would be fine as pages wouldn't look too empty. Or, It's just a headline and the editor, where a single page would feel like "richer" documentation (for the lack of a better word)

It's definitely going to mostly just be a headline and an editor. At most, I'll probably include a sentence or two on the top of the editor, but not explain what the code is doing (comments can do that if needed)

P.S Glad you made this suggestion, as it reminded me of 30 seconds of CSS, which is essentially what this is trying to do now that I think about it. And they also refer to their stuff as snippets so definitely the naming here is correct! :smile: :tada: 

---

So I'll go ahead and do the suggestions and move into a single file once I get the chance :smile: 

Thanks y'all! :smile: :tada:

src/snippets/switch.md
torch2424: This is intended but I'll leave a comment to specify that :smile:

src/snippets/class.md
torch2424: Ah, so I'm a little lost on what you mean here regarding the "good last case in the HTML" :upside_down_face: . But, I'm down for whatever suggestion helps show off the syntax / pattern for using classes :smile: :joy:

src/snippets/extending-classes.md
torch2424: Same as above :smile: :+1:

src/snippets/handling-null.md
torch2424: Ah yeah so I don't know a good way to express that in a snippet. Perhaps we can leave a comment (in markdown) as a general idea? 

Ideally, we would have a double equals vs. triple equals snippet that shows off this behavior and link to it from this example :smile:

torch2424:
@dcodeIO @MaxGraey Made all requested changes! This is now good to go! :smile: :+1: 

Screenshot from 2020-06-11 17-42-32

torch2424:
Updated to latest master @dcodeIO @MaxGraey :smile: 

Screenshot from 2020-06-23 14-29-48

src/snippets.md
```diff
@@ -0,0 +1,294 @@
...
+  console.log(`switchSurprise(3) : ${exports.switchSurprise(3)}`);
+  console.log(`switchSurprise(4) : ${exports.switchSurprise(4)}`);
+  console.log(`switchSurprise(57) : ${exports.switchSurprise(57)}`);
+});
+</script>
+```
+
+## Ternary
+
+Using Ternary in AssemblyScript
```
aminya:
```suggestion
## Ternary if-else

Using Ternary if-else in AssemblyScript
```

Ternary can have [a lot of meanings](https://en.wikipedia.org/wiki/Ternary_operation#:~:text=A%20ternary%20operation%20on%20a%20set%20A%20takes,operator%20is%20an%20operator%20that%20takes%20three%20arguments.). It is better to be exact here 😁

src/snippets.md
```diff
@@ -0,0 +1,294 @@
...
+  } else {
+  console.log("Could not add -1");
+  }
+});
+</script>
+```
+
+## Switch
+
+Using `switch` statements in AssemblyScript.
```
aminya:
```suggestion
## Switch Case

Using `switch case` statements in AssemblyScript.
```

src/snippets.md
```diff
@@ -0,0 +1,294 @@
...
+.then(({ exports }) => {
+  console.log(`getStaticProp: ${exports.getStaticProp()}`);
+  console.log(`overloadAdd: ${exports.overloadAdd(24)}`);
+})
+</script>
+```
+
+## Handling Null
+
+Handling Null as a union type for optional return values, and simulating try/catch for errors in AssemblyScript
```
aminya:
```suggestion
Handling Null as a union type for optional return values and simulating `try/catch` for errors in AssemblyScript
```

src/snippets.md
```diff
@@ -0,0 +1,294 @@
...
+    console.log(exports.staticAdd(1, 2));
+    console.log(exports.instanceTwo());
+    console.log(exports.instanceSub(3.0, 1.0));
+  })
+</script>
+```
+
+## Extending Classes
+
+Extending classes, and general Object Orientated Programming (OOP) in AssemblyScript
```
aminya:
```suggestion
Extending classes and general Object Orientated Programming (OOP) in AssemblyScript
```

src/snippets.md
```diff
@@ -0,0 +1,294 @@
+# Snippets
+
+Snippets are small, copy-pastable AssemblyScript examples meant to show off common syntax and patterns when writing AssemblyScript.
+
+## Class
+
+A snippet on using Classes in AssemblyScript, and assigning instance vs. static properties.
```
aminya:
```suggestion
An AssemblyScript snippet on using Classes, their instantiation, and their static properties.
```

aminya (approved):

src/snippets.md
```diff
@@ -0,0 +1,294 @@
...
+    if (typeof arg == 'object') {
+      str += `${JSON.stringify(arg, null, 2)}<br/>`;
+    } else {
+      str += `${arg}<br/>`;
+    }
+  });
+  document.body.innerHTML += `<div>Log: ${str}</div>`;
+}
+
+loader.instantiate(module_wasm, { /* imports */ })
```
aminya, line 181: Adding an example instead of `/* imports */`?

src/snippets.md
```diff
@@ -0,0 +1,294 @@
...
+  }
+});
+</script>
+```
+
+## Switch
+
+Using `switch` statements in AssemblyScript.
+
+**NOTE:** Switch conditions currently implicitly convert to u32, i.e. switching over strings or similar is not yet supported. If using i32s, a more technical detail would be that using values greater than or equal to zero has better codegen, because br_tables are unsigned, so a -1 for example would not tablify well with otherwise positive values (it does work, though, but code is not optimal).
```
aminya:
```suggestion
**NOTE:** Currently, the switch conditions (case values) are implicitly converted to u32, i.e. switching over strings or similar is not yet supported.
```

I don't understand this part 🙄, and I am not sure if these details are necessary here:
```
 If using i32s in the switch conditions, a more technical detail would be that using values greater than or equal to zero has better codegen, because br_tables are unsigned, so a -1, for example, would not tablify well with otherwise positive values (it does work, though, but code is not optimal).
```

src/snippets.md
```diff
@@ -0,0 +1,294 @@
...
+  // To show both syntaxes are supported here.
+  switch (a) {
+    case 1:
+      response = 100;
+      break;
+    // Cases can also use braces
+    case 2: {
+      response = 200;
+      break;
+    }
```
aminya:
```suggestion
    case 2: {   // Cases can also use braces
      response = 200;
      break;
    }
```

src/snippets.md
```diff
@@ -0,0 +1,294 @@
...
+}
+
+#!html
+<script>
+const log = console.log;
+console.log = (...args) => {
+  log(...args);
+  let str = '';
+  args.forEach(arg => {
+    if (typeof arg == 'object') {
```
aminya:
```suggestion
    if (typeof arg === 'object') {
```

src/snippets.md
```diff
@@ -0,0 +1,294 @@
...
+      str += `${JSON.stringify(arg, null, 2)}<br/>`;
+    } else {
+      str += `${arg}<br/>`;
+    }
+  });
+  document.body.innerHTML += `<div>Log: ${str}</div>`;
+}
+
+loader.instantiate(module_wasm, { /* imports */ })
```
aminya, line 124: Same as https://github.com/AssemblyScript/website/pull/2/files#r464214621?

src/snippets.md
```diff
@@ -0,0 +1,294 @@
...
+
+**NOTE:** Switch conditions currently implicitly convert to u32, i.e. switching over strings or similar is not yet supported. If using i32s, a more technical detail would be that using values greater than or equal to zero has better codegen, because br_tables are unsigned, so a -1 for example would not tablify well with otherwise positive values (it does work, though, but code is not optimal).
+
+```editor
+#!runtime=half
+export function switchSurprise(a: i32): i32 {
+  let response = -1;
+
+  // Using a mix of braces and not using braces
+  // To show both syntaxes are supported here.
```
aminya:
```suggestion
  // to show that both syntaxes are supported here.
```

src/snippets.md
dcodeIO: Hmm, I suggested this in https://github.com/AssemblyScript/website/pull/2#discussion_r436265041, but if you say it's not helping then I'm fine with keeping this simple.

dcodeIO:
Totally forgot about the snippets, sorry :) Took the liberty to include @aminya's suggestions since these all looked good.

dcodeIO:
Thanks guys!

torch2424:
Yay!!!

> Totally forgot about the snippets, sorry :)

No worries! I didn't know if you were busy, or forgot. But I know you are often being reached out by everyone to do random things, so I didn't want to bring it up and figured the time would come eventually :joy: 

> Took the liberty to include @aminya's suggestions since these all looked good.

Huge thank you @dcodeIO for doing all the awesome suggestions by @aminya :smile: 


--


Glad this is merged! I know I had a list somewhere of other random snippets I wanted to add. So I'll implement those when I have the chance :smile: :+1:

PR #4, opened 2020-06-11
Add a quirks section

dcodeIO:
As discussed on Slack recently, this PR restructures what previously were "limitations" to a more general "quirks" section, where we can mention more aspects like null checking on fields. Moves the FAQ to its own page again so one can skip it more naturally and the basics page doesn't get too overwhelming. Also moves "built-with-assemblyscript" to examples, where it fits better.

cc @trusktr


dcodeIO:
Merging since no objections :)

PR #21, opened 2020-10-05
Say "Unicode code point" instead of "UTF-8 code point".

sunfishcode:
This avoids ambiguity since code points are a Unicode-level concept, and
aren't specific to UTF-8.


dcodeIO:
Thanks!

PR #32, opened 2020-11-20
update quirks on nullability

trusktr:


src/basics.md
```diff
@@ -128,6 +128,26 @@ function doSomething(foo: Foo): void {
 }
 ```
 
+Nullability checks are currently not supported in OR statements. Instead of writing
+
+```ts
+function foo(someVar: Thing | null) {
+  someVar = someVar || new Thing
+  
+  // AS still sees someVar as 'Thing | null' here
```
dcodeIO, line 137: Hmm this one seems like a bug. Should perhaps fix that instead.

PR #45, opened 2021-01-29
Added Wasmer to the functionality list

syrusakbary:
Great work on AssemblyScript 0.18!

I just read the announcement pointing to the [status page](https://www.assemblyscript.org/status.html) and realized that it might be a good idea to list [Wasmer](https://wasmer.io) in the [WebAssembly Features section](https://www.assemblyscript.org/status.html#webassembly-features) (as more standalone runtimes are already listed).

Let me know if this makes sense or if any amendment is needed!
Keep up the good work 👏


dcodeIO:
Hi Syrus! I have been wondering about the relation of Wasmtime and Wasmer for a while, perhaps you can help me to understand better. Is Wasmer a soft or a hard fork of Wasmtime, i.e. would it be reasonable to expect both to have a similar feature set as soon as the other supports it? I am mostly interested in the raw capability to run an AssemblyScript module with certain Wasm features enabled, not so much in anything beyond that. Also, do you have a reference somewhere, in your documentation perhaps, outlining Wasm features and their implementation status?

syrusakbary:
Hi @dcodeIO, thanks for the question. This is a great opportunity to help on a better understanding between VMs.

Wasmer is neither a hard fork or a soft fork of wasmtime, neither wasmtime is a hard or soft work of Wasmer.
We are two different runtimes with two different implementations, driven by different teams. You may want to check our [Attributions](https://github.com/wasmerio/wasmer/blob/master/ATTRIBUTIONS.md) if you want to dig further regarding our codebase!

Because of that, the feature set of things supported by Wasmer is independent of the supported by other VMs.
For example, when [compared to wasmtime](https://wasmer.io/wasmer-vs-wasmtime), Wasmer can run multiple compilers (LLVM, Cranelift and Singlepass), run in headless mode, with a Native engine (similar to Lucet) and in many more programming languages.

> would it be reasonable to expect both to have a similar feature set as soon as the other supports it?

Not necessary. In the same way other VMs (v8, spidermonkey, wasmtime or WAVM) have their own roadmap, we have ours. We aim to support all WebAssembly features in Phase 4 onwards, but we do it in our own roadmap and schedule.

For example, we supported threads and [SIMD](https://medium.com/wasmer/webassembly-and-simd-13badb9bf1a8) before any other standalone WebAssembly VM (except for WAVM), but we are one of the last ones implementing Reference Types.

> do you have a reference somewhere, in your documentation perhaps, outlining Wasm features and their implementation status?

Yes, we do :).

Here's a list of features supported by Wasmer in our docs:
https://docs.wasmer.io/ecosystem/wasmer/wasmer-features#support-of-features-by-compiler

And also the official comparison of WebAssembly features by runtime (we try to keep it up to date):
https://webassembly.org/roadmap/

Let me know if this clarifies your questions!

dcodeIO:
Cool, thanks for the thorough explanation!

src/status.md
```diff
@@ -58,7 +58,8 @@ Some crucial language features rely on [future WebAssembly functionality](https:
 <Fi/> Firefox &nbsp;
 <Sa/> Safari &nbsp;
 <No/> Node.js &nbsp;
-<Wa/> Wasmtime (<sup>1</sup> native i64 support)
+<Wa/> Wasmtime &nbsp;
+<Wasmer/> Wasmer (<sup>1</sup> native i64 support)
```
dcodeIO, line 62:
```suggestion
<Ws/> [Wasmer](https://docs.wasmer.io/ecosystem/wasmer/wasmer-features#support-of-features-by-compiler) (<sup>1</sup> native i64 support)
```

src/.vuepress/components/Wasmer.vue
```diff
@@ -0,0 +1,3 @@
+<template>
```
dcodeIO, line 1: Let's name this file `Ws.vue` to keep the table using it a bit narrower. I may then later rename `Wa` to `Wt` to avoid confusion.

dcodeIO (commented):
LGTM with the component renamed and the nice-to-have feature link added :)

dcodeIO:
Since just a couple nits, I guess I can as well just merge and make the changes myself pretty quick afterwards. Thanks! :)

PR #47, opened 2021-01-31
Add notes on using manual memory management with GC

dcodeIO:
As suggested by Max, this adds a few more details on using manual memory management in parallel to GC.


MaxGraey (approved):
LGTM

PR #89, opened 2021-07-25
Fix a typo in notice

y-yagi:


dcodeIO:
Oh, something must have gone wrong there. We might have used AssemblyScript on a WASI host to generate this message.

dcodeIO:
Oh, something must have gone wrong there. We might have used AssemblyScript on a WASI host to generate this message. Or we might have attempted to roundtrip the message through WebAssembly using JavaScript. Or the message was generated by C#. Perhaps also Java. Can't really tell. Need to investigate. Seems to be silent data corruption.

dcodeIO:
It will take me a while to find out, these issues are really hard to debug. Might be that our entire language is broken. Going to close this issue to preserve the error for debugging purposes. Thanks for reporting, you might just have found a bug that can break the entire Web Platform!

PR #90, opened 2021-07-26
improve the banner to be more encouraging for people to get involved

trusktr:


torch2424:
I already said this in our core team chat, but just wanted to post here in case anyone comes across this 😄 

I think our best bet here is to take down the notice, and then start from scratch with whatever messaging we want to decide upon. Because it shows that the message is in flux, and less people (Especially people who depend on AssemblyScript already) will see a message "we as a team" don't necessarily agree with it.

Personally, I think we should be stepping back and starting the message from step 1. Let's decide what our goals with a message is, and then probably what would be the best way to communicate that?

But! The message was posted by @dcodeIO , who is the project lead, author, and BDFL of the project: https://github.com/AssemblyScript/core#people . Who technically per the project governance totally can do so, but I just want to throw out there that's my opinion on the best next step 👍🏾 

Thanks!

trusktr:
@torch2424 Here's a PR for that: 

https://github.com/AssemblyScript/website/pull/91

torch2424:
Thank you very much @trusktr I hella appreciate that! 😀🎉

PR #105, opened 2021-10-01
fix usize type description

mathe42:
I think the value in the table is wrong as `isize = usize` seems wrong...


dcodeIO:
This can be confusing, yet there is no `u32` (respectively `u64`) in Wasm. There is just `i32` with different sets of instructions to perform unsigned operations, for example `i32.div_s` vs `i32.div_u`. As such the WebAssembly type backing `usize` and `isize` is in fact the same.

mathe42:
OK LOL

PR #122, opened 2022-01-07
Add as-proto library

piotr-oles:
I made a small library to encode/decode protobuf messages. I wouldn't use it on production, but I guess it's worth sharing with the community :)

https://github.com/piotr-oles/as-proto


piotr-oles:
gentle ping @dcodeIO :)

PR #124, opened 2022-01-26
Update documentation for 0.20

dcodeIO:
Initial documentation updates accompanying https://github.com/AssemblyScript/assemblyscript/pull/2157.

TODO:

- [x] Update asinit to utilize host bindings and reflect the change in documentation
- [x] Check that there's no more `require`
- [x] Update the editor to use the new ESM distribution files
- [x] Revise examples

To preview the updated documentation in a more convenient form:

```sh
git clone --branch esm https://github.com/AssemblyScript/website.git
cd website
npm install
npm start
```

The website will then be served at http://localhost:8080/


src/introduction.md
```diff
@@ -21,20 +20,15 @@ export function fib(n: i32): i32 {
...
-
 ```sh
-asc fib.ts --binaryFile fib.wasm --optimize
+asc fib.ts --out fib.wasm --optimize
 ```
 
-As such, it differs from running dynamically typed JavaScript (just in time) in that it instead [statically compiles](./compiler.md) to a strictly typed WebAssembly binary **ahead of time**, quite similar to what a traditional compiler would do. However, since it is deliberately designed to be very similar to JavaScript, it has the potential to integrate seamlessly with existing Web Platform concepts to produce lean and mean WebAssembly binaries, while also making it almost natural to use for those who are already familiar with writing code for the Web.
+It is very similiar to **TypeScript but with WebAssembly types**, has some constraints due to compiling **strictly typed** code **ahead of time**, but also some additions originating in WebAssembly's feature set. Its close relation to JavaScript makes it a familiar choice for developers who are already used to writing code for the Web, and also has the potential to integrate seamlessly with existing Web Platform concepts to produce lean and mean WebAssembly modules.
```
romdotdog: It may be worth noting that AssemblyScript can't compile TypeScript.

src/introduction.md
```diff
@@ -46,22 +40,53 @@ For comparison, the following C code is roughly equivalent:
...
-## Setting realistic expectations
+Both perspectives can be mixed depending on whether raw performance with WebAssembly instructions or idiomatic concepts with the managed standard library are desirable to accomplish an individual task.
+
+## Frequently asked questions
+
+::: tip Does AssemblyScript involve an interpreter, or a "VM in a VM"?
+No, AssemblyScript compiles to WebAssembly bytecode directly, statically, ahead-of-time.
+:::
+
+::: tip What are the differences between AssemblyScript and Typescript?
```
romdotdog:
```suggestion
::: tip What are the differences between AssemblyScript and TypeScript?
```

src/introduction.md
```diff
@@ -46,22 +40,53 @@ For comparison, the following C code is roughly equivalent:
...
+:::
 
-[Strictly typed](./basics.md#strictness) AssemblyScript differs a bit from idiomatic TypeScript, which even though typed, attempts to describe all the dynamic features of JavaScript, many of which cannot be efficiently compiled ahead of time. Not a blocker, but takes a bit to get used to.
+::: tip Can AssemblyScript be used outside of the browser?
+Absolutely! AssemblyScript modules are self-contained and run anywhere where WebAssembly is supported.
+:::
 
-Also, we are patiently waiting for [future WebAssembly features](./status.md) (marked as 🦄 throughout the documentation) to become available for us to use. In particular, we are looking forward to proposals that promise to improve WebAssembly's integration with the Web Platform, e.g. to more naturally and efficiently share strings, arrays and objects with JavaScript.
+::: tip Why the strange name?
+AssemblyScript is to Assembly as JavaScript is to Java. Gotta stick to the theme.
```
romdotdog:
```suggestion
AssemblyScript is to WebAssembly as JavaScript is to Java. Gotta stick to the theme.
```
Overall, this section was both amusing and confusing. JavaScript, in my opinion, doesn't bear much of a relation to Java, while AssemblyScript bears a relation to (WebAssembly?)

Or is the joke that AssemblyScript is not very related to Assembly in general?

src/getting-started.md
```diff
@@ -0,0 +1,101 @@
...
+Doing so will emit the compiled binaries, bindings and definition files to the `build/` directory.
+
+The generated test case in `tests/index.js` can be executed with:
+
+```sh
+npm test
+```
+
+Once built, the directory contains all the bits to use the module like any other modern Node.js
+module (ESM format), e.g. when it is installed as a dependency of another module:
```
romdotdog: Could be better phrased as
```
Once built, the directory contains all the bits to use the module like any other modern Node.js ESM
module:
```

src/getting-started.md
```diff
@@ -0,0 +1,101 @@
...
+
+## The journey ahead
+
+So far, so good! Now it is time to start editing the project of course, which typically involves:
+
+* Editing and adding source files within the `assembly/` directory and updating the tests in `tests/`.
+
+* Tweaking [compiler options](./compiler.md) in [`asconfig.json`](./compiler.md#asconfig-json) to fit your project's needs.
+
+* Realizing that AssemblyScript is not quite TypeScript and that WebAssembly still has a way to go 🙂
```
romdotdog:
```suggestion
* Realizing that WebAssembly still has a ways to go 🙂
```
I'm not sure that AssemblyScript's goal is to be a 100% TypeScript to WebAssembly compiler. If we make that clear throughout the docs then people wouldn't have the impression that AssemblyScript has TypeScript's paradigms.

src/compiler.md
```diff
@@ -86,28 +54,16 @@ Typical output formats are WebAssembly binary \(.wasm\) and/or text format \(.wa
 --outFile, -o         Specifies the output file. File extension indicates format.
```
romdotdog, line 56: Move `--noEmit` here?

src/compiler.md
```diff
@@ -191,27 +152,348 @@ Other options include those forwarded to Binaryen and various flags useful in ce
...
+Externref       | By reference | Using reference-types.
+**String**      | Copy         |
+**ArrayBuffer** | Copy         |
+**TypedArray**  | Copy         | Any `Int8Array`, `Float64Array` etc.
+**Array**       | Copy         | Any `Array<T>`
+**StaticArray** | Copy         | Any `StaticArray<T>`
+**Object**      | Copy         | If a plain object. That is: Has no constructor or non-public fields.
+**Object**      | By reference | If not a plain object. Passed as an opaque reference counted pointer.
+
+Note the two different strategies used for **Object**: In some situations, say when calling a Web API, it may be preferrable to copy the object as a whole, field by field, which is the strategy chosen for plain objects with no constructor or non-public fields:
```
romdotdog:
```suggestion
Note the two different strategies used for **Object**: In some situations, say when calling a Web API, it may be preferable to copy the object as a whole, field by field, which is the strategy chosen for plain objects with no constructor or non-public fields:
```

src/compiler.md
```diff
@@ -191,27 +152,348 @@ Other options include those forwarded to Binaryen and various flags useful in ce
...
+  return target.field;
+}
+```
+
+Also note that exporting an entire `class` has no effect at the module boundary (yet), and it is instead recommended to expose only the needed functionality as shown in the example above. Supported elements at the boundary are globals, functions and enums.
+
+
+## Debugging
+
+The debugging workflow is similar to debugging JavaScript (it's one engine) and the compiler provides various options to set up additional WebAssembly-specific debug information. Note that any sort of optimization should be disabled in debug builds.
```
romdotdog: The "it's one engine" confused me a bit here.

src/concepts.md
```diff
@@ -0,0 +1,360 @@
...
+  var b = a + 1
+  return b
+}
+```
+
+There are no union types, but a similar effect can be achieved with generics:
+
+```ts
+// 😢
+function foo(a: i32 | string): void {
```
romdotdog, line 39: It is a matter of opinion, but I would insert `// ...` in the function bodies.

src/concepts.md
```diff
@@ -0,0 +1,360 @@
...
+a.set("prop", "hello world")
+
+// 😊
+class A {
+  constructor(public prop: string) {}
+}
+var a = new A("hello world")
+```
+
+And nullability checks are limited to locals to guarantee soundness where TypeScript [would](https://www.typescriptlang.org/play?#code/G4QwTgBAZg9jBcEDOAXMBLAdgcwgHwkwFcAbEiAXggCIRqBuAKEfSggApYYBKCAb0YRkAUxQAxOADlSJdtyZCuAOhLCcKABb0IAeh2EYEYWDAwwjAL7MoRTAGMU6GJhHipMuf0HQ4lQjKYrIA) not [diagnose](https://www.typescriptlang.org/play/index.html#code/MYGwhgzhAEBiD29oG8BQ0PQvAtgUwBcALASwDsBzALiwICdyLoAfaMgVxBGgF43OQAblQBfVKgBm7MsAIl4ZaABN4AZVyFSlABQTENBPACUKdJhITouxADps+YoxNpMr5Wo2PKAURAQ81sbCbph68HaeWhQ2IHiUxILQAPRJWETwnErQeHR08HQANNAARuwE7ngQZADkBGYYYmKS0rLyiirqDlG+-oEGiM710GERXYy8-FzCIkA) a [problem](https://www.typescriptlang.org/play/index.html#code/MYGwhgzhAEBiD29oG8BQ0PQPoXgWwFMAXACwEsA7AcwC5oIiAnSq6AH2goFcQRoBeaACISBXvCEBudJirF6+YuWoAKAJR0Gzau048+aTEegh5jeYNJkIAOhyKr1accxXb9wo9aDuvZy+hzIi5GCkDifwwAX1QY1AAzLgpgIjJ4MIATeABlB2UqFXjEOgR4NRQZDDJ46ELEG1xPfPLDAKL4BryWG1NqUkloAHpB+hJ4HgzoAkZGeEYAGmgAIy4iaCyCCAoAciJK6BiYoA):
```
romdotdog:
```suggestion
And nullability checks are limited to locals to guarantee soundness where TypeScript [would not](https://www.typescriptlang.org/play?#code/G4QwTgBAZg9jBcEDOAXMBLAdgcwgHwkwFcAbEiAXggCIRqBuAKEfSggApYYBKCAb0YRkAUxQAxOADlSJdtyZCuAOhLCcKABb0IAeh2EYEYWDAwwjAL7MoRTAGMU6GJhHipMuf0HQ4lQjKYrIA) [diagnose](https://www.typescriptlang.org/play/index.html#code/MYGwhgzhAEBiD29oG8BQ0PQvAtgUwBcALASwDsBzALiwICdyLoAfaMgVxBGgF43OQAblQBfVKgBm7MsAIl4ZaABN4AZVyFSlABQTENBPACUKdJhITouxADps+YoxNpMr5Wo2PKAURAQ81sbCbph68HaeWhQ2IHiUxILQAPRJWETwnErQeHR08HQANNAARuwE7ngQZADkBGYYYmKS0rLyiirqDlG+-oEGiM710GERXYy8-FzCIkA) [a problem](https://www.typescriptlang.org/play/index.html#code/MYGwhgzhAEBiD29oG8BQ0PQPoXgWwFMAXACwEsA7AcwC5oIiAnSq6AH2goFcQRoBeaACISBXvCEBudJirF6+YuWoAKAJR0Gzau048+aTEegh5jeYNJkIAOhyKr1accxXb9wo9aDuvZy+hzIi5GCkDifwwAX1QY1AAzLgpgIjJ4MIATeABlB2UqFXjEOgR4NRQZDDJ46ELEG1xPfPLDAKL4BryWG1NqUkloAHpB+hJ4HgzoAkZGeEYAGmgAIy4iaCyCCAoAciJK6BiYoA):
```

src/concepts.md
```diff
@@ -0,0 +1,360 @@
...
+
+logInteger(42)
+```
+
+Ambient declarations in an AssemblyScript file will yield a WebAssembly module import, using the internal path of the file, without file extension, as the module name (here: `assembly/env`), and the name of the declared element as the module element (here `logInteger`). In the example above, the import can be fulfilled by providing the following imports object upon instantiation:
+
+```js
+WebAssembly.instantiateStreaming(fetch(...), {
+  "assembly/env": {
+    "logInteger"(i) { console.log("logInteger: " + i) }
```
romdotdog:
```suggestion
    logInteger(i) { console.log("logInteger: " + i) }
```

src/getting-started.md
```diff
@@ -0,0 +1,101 @@
+
+# Getting started
+
+Paving the way to your first AssemblyScript module.
+
+## Setting up a new project
+
+Make sure that a [recent version of Node.js](https://nodejs.org) and its package manager **npm** \(comes with Node.js\) are installed, then switch to a new directory and initialize a new Node.js module as usual:
```
romdotdog:
```suggestion
Make sure that a [recent version of Node.js](https://nodejs.org) and its package manager **npm** \(that comes with Node.js\) are installed, then switch to a new directory and initialize a new Node.js module as usual:
```
Seems a bit more eloquent.

src/getting-started.md
```diff
@@ -0,0 +1,101 @@
...
+  ./package.json
+  Package info containing the necessary commands to compile to WebAssembly.
+
+  ./tests/index.js
+  Stater test to check that the module is functioning.
+
+  ./index.html
+  Starter HTML file that loads the module in a browser.
+```
+
```
romdotdog, line 56: Note here that `asinit` has a few other options that can be listed using `asinit --help`?

src/getting-started.md
```diff
@@ -0,0 +1,101 @@
...
+So far, so good! Now it is time to start editing the project of course, which typically involves:
+
+* Editing and adding source files within the `assembly/` directory and updating the tests in `tests/`.
+
+* Tweaking [compiler options](./compiler.md) in [`asconfig.json`](./compiler.md#asconfig-json) to fit your project's needs.
+
+* Realizing that AssemblyScript is not quite TypeScript and that WebAssembly still has a way to go 🙂
+
+But that's it already for a quick start. Read on to [learn more](/basics.md)!
```
romdotdog: Dead link.

src/getting-started.md
```diff
@@ -0,0 +1,101 @@
...
+
+The generated `index.html` shows how the module can be used on the Web. A web server serving
+the module directory, defaulting to display `index.html`, can be started with:
+
+```sh
+npm start
+```
+
+Note that not all of the files may be required depending on the use case, and it is safe
+to delete what's not needed. If anything goes wrong, `asinit` can be executed again, then
```
romdotdog:
```suggestion
to delete what's not needed. If anything goes wrong, `asinit` can be executed again, which would
```

src/getting-started.md
```diff
@@ -0,0 +1,101 @@
...
+restoring the deleted default files while keeping already edited ones.
```
romdotdog:
```suggestion
restore the deleted default files while keeping already edited ones.
```

src/runtime.md
```diff
@@ -4,160 +4,114 @@ description: Because nobody enjoys being bullied by malloc, free and friends.
...
+  --exportRuntime       Exports the runtime helpers (__new, __collect etc.).
+```
 
-::: tip
-Previous versions of the compiler \(pre 0.7\) did not include any runtime functionality by default but instead required `import`ing an allocator and potentially an experimental tracing collector. This is not supported anymore by recent versions of the compiler.
-:::
+The default `incremental` runtime provides the full package recommended in most use cases. The `minimal` runtime is a stripped down variant (no shadow stack, no heuristic, simpler algorithm, smaller and more efficient) that is not automated and requires calling `__collect` externally at appropriate times (when the execution stack in unwound), whereas the `stub` runtime does not provide a garbage collector at all and never frees (simple bump allocation, extremely small).
```
romdotdog:
```suggestion
The default `incremental` runtime provides the full package recommended in most use cases. The `minimal` runtime is a stripped down variant (no shadow stack, no heuristic, simpler algorithm, smaller and more efficient) that is not automated and requires calling `__collect` externally at appropriate times (when the execution stack is unwound), whereas the `stub` runtime does not provide a garbage collector at all and never frees (simple bump allocation, extremely small).
```
Elaborate on what an unwound execution stack is exactly?

src/runtime.md
```diff
@@ -4,160 +4,114 @@ description: Because nobody enjoys being bullied by malloc, free and friends.
...
+It is typically not necessary to invoke the runtime interface manually since generated bindings take care of all the internals. In environments where bindings are not yet available, however, the runtime interface can be utilized to provide the necessary integration.
 
 * ```ts
   function __new(size: usize, id: u32): usize
   ```
-  Dynamically allocates an object represented by the specified id of at least the given size in bytes and returns its address. Alignment is guaranteed to be 16 bytes to fit up to v128 values naturally. Does not zero memory.
-
-The [loader](./loader.md) provides some additional functionality for convenience, like `__newString`.
+  Allocates a new garbage collected instance of the object represented by the specified class id, of at least the specified size. Returns the pointer to the object, typically pointing at the object's data respectively its first field.
```
romdotdog:
```
typically pointing at the object's data respectively its first field.
```
This doesn't make sense.

src/runtime.md
```diff
@@ -4,160 +4,114 @@ description: Because nobody enjoys being bullied by malloc, free and friends.
...
 
-### Managing lifetimes
+* ```ts
+  function __pin(ptr: usize): usize
+  ```
+  Pins the object pointed to by `ptr` externally so it and its directly reachable members and indirectly reachable objects do not become garbage collected. Note that the same object cannot be pinned more than once.
```
romdotdog: > its directly reachable members and indirectly reachable objects 

To be honest, I have no idea what this means.

src/runtime.md
```diff
@@ -4,160 +4,114 @@ description: Because nobody enjoys being bullied by malloc, free and friends.
...
 
-The concept is simple: If a reference to an object is established, the reference count of the object is increased by 1 \(a reference is retained\), and when a reference to an object is deleted, the reference count of the object is decreased by 1 \(a reference is released\). If the reference count of an object reaches 0, it is considered for collection and its memory ultimately returned to the memory manager for reuse. The relevant interface is:
+  An external object that is not referenced from within WebAssembly must be pinned whenever an allocation might happen in between its allocation and passing it to WebAssembly. If not pinned, the allocation may trigger the garbage collector to step, which would prematurely collect the object and lead to undefined behavior.
```
romdotdog:
```suggestion
  An external object that is not referenced from within WebAssembly must be pinned whenever an allocation might happen in between allocating it and passing it to WebAssembly. If not pinned, the allocation may trigger the garbage collector to step, which would prematurely collect the object and lead to undefined behavior.
```

src/runtime.md
```diff
@@ -4,160 +4,114 @@ description: Because nobody enjoys being bullied by malloc, free and friends.
...
 
-By default, the full and half runtime will automatically try to collect cyclic garbage when memory must be grown. This behavior can be disabled by setting `gc.auto = false` in performance critical code. Likewise, if there is a good opportunity to collect cyclic garbage at a given point in time, like if the application is idle, `gc.collect()` can be called to force a full garbage collection cycle. Protip: If no cyclic structures are used, no garbage must be collected.
+## Memory layout
 
-## Implementation
+Besides partitioning memory into static data (starts at `0`), the stack (starts at `__data_end`) and the heap (starts at `__heap_base`), in this order, any kind of managed object in AssemblyScript utilizes a managed header for the runtime to operate on:
```
romdotdog: The syntax of this sentence took me a while to understand. Rewrite?

src/runtime.md
```diff
@@ -4,160 +4,114 @@ description: Because nobody enjoys being bullied by malloc, free and friends.
...
+| mmInfo   |    -20 | usize | Memory manager info
+| gcInfo   |    -16 | usize | Garbage collector info
+| gcInfo2  |    -12 | usize | Garbage collector info
+| rtId     |     -8 | u32   | Unique id of the concrete class
+| rtSize   |     -4 | u32   | Size of the data following the header
+|          |        |       |
+|          |      0 |       | Payload starts here
 
-Both the TLSF memory manager and the concept of reference counting are a good fit for _predictable performance_ scenarios. Remember: WebAssembly is about exactly that. It is not necessarily the fastest choice in every possible scenario, though.
+References to an object always point at the start of the payload, with the header sitting in the 20 bytes before. Null references are just the value `0`. When working with an AssemblyScript module externally, knowing the memory layout can be helpful to for example obtain an object's class id or size. Invoking `__new` automatically prepends a managed header and registers the object with the GC, using the provided class id for `rtId` and the provided size for `rtSize`.
```
romdotdog:
```suggestion
References to an object always point at the start of the payload, with the header beginning 20 bytes before. Null references are just the value `0`. When working with an AssemblyScript module externally, knowing the memory layout can be helpful to for example obtain an object's class id or size. Invoking `__new` automatically prepends a managed header and registers the object with the GC, using the provided class id for `rtId` and the provided size for `rtSize`.
```

src/runtime.md
```diff
@@ -4,160 +4,114 @@ description: Because nobody enjoys being bullied by malloc, free and friends.
...
 
-It is also likely that our implementations are not as optimized yet as ultimately possible. If you are smarter than us, please let us know of your thoughts and findings.
+### Class layout
 
-## Internals
+Class fields are layed out similar to C structs, sequentially and without packing. Each field is aligned to its type's native alignment, potentially leaving padding in between. If a class has the `@unmanaged` decorator, it effectively only describes a region of memory as if it was a struct, does not utilize a managed header, is not garbage collected, and can be used with `heap.free`. Managed classes with a managed header cannot be manually freed, and managed and unmanaged classes cannot be mixed (e.g. extend from each other).
```
romdotdog:
```suggestion
Class fields are laid out similarly to C structs: sequentially and without packing. Each field is aligned to its type's native alignment, potentially leaving padding in between. If a class has the `@unmanaged` decorator, it effectively only describes a region of memory as if it were a struct that does not utilize a managed header, is not garbage collected, and can be used with `heap.free`. Managed classes with a managed header cannot be manually freed, and managed and unmanaged classes cannot be mixed (e.g. extend from each other).
```
https://www.grammarly.com/blog/conditional-sentences-was-instead-of-were/

src/runtime.md
```diff
@@ -4,160 +4,114 @@ description: Because nobody enjoys being bullied by malloc, free and friends.
...
 
-If you are interested in the inner workings, the internal APIs are explained in [the runtime's README file](https://github.com/AssemblyScript/assemblyscript/tree/main/std/assembly/rt) and of course in its sources - feel free to take a look!
+Standard library data types use the following layouts:
 
-#### Runtime type information \(RTTI\)
+Class           | Description
+----------------|-------------
+ArrayBuffer     | Buffers always use class id `0`, with their untyped data as the payload.
+String          | Strings always use class id `1`, with their 16-bit char codes (UTF-16 code units, allowing isolated surrogates like JS) as the payload. For example, if `rtSize` is `8`, the string's `.length` is `4`.
+TypedArray      | Typed arrays are objects composed of `buffer` (reference to the viewed `ArrayBuffer`), `dataStart` (start pointer into `buffer`) and `byteLength` fields, in this order. The respective id is picked sequentially and not predetermined.
```
romdotdog:
```suggestion
TypedArray      | Typed arrays are objects composed of `buffer` (the reference to the viewed `ArrayBuffer`), `dataStart` (the start pointer into `buffer`) and `byteLength` fields, in this order. The respective id is picked sequentially and not predetermined.
```

src/runtime.md
```diff
@@ -4,160 +4,114 @@ description: Because nobody enjoys being bullied by malloc, free and friends.
...
+Array\<T>       | Normal arrays use the same layout as the typed arrays, with an additional mutable `length` field coming last.
```
romdotdog:
```suggestion
Array\<T>       | Normal arrays use the same layout as typed arrays, with an additional mutable `length` field coming last.
```

src/runtime.md
```diff
@@ -4,160 +4,114 @@ description: Because nobody enjoys being bullied by malloc, free and friends.
...
+StaticArray\<T> | Static arrays do not need indirection due to not being resizable and have their data right in the payload, aligned according to `T`. Can be thought of as a typed buffer.
```
romdotdog:
```suggestion
StaticArray\<T> | Static arrays do not need indirection due to not being resizable, and they have their data right in the payload, aligned according to `T`. Can be thought of as a typed buffer.
```

src/runtime.md
```diff
@@ -4,160 +4,114 @@ description: Because nobody enjoys being bullied by malloc, free and friends.
...
+Map/Set/...     | Other collection use more complex layouts. Please refer to their respective sources.
 
-Every module using managed objects contains a memory segment with basic type information, that [the loader](./loader.md) for example uses when allocating new arrays. Internally, RTTI is used to perform dynamic `instanceof` checks and to determine whether a class is inherently acyclic. The memory offset of RTTI can be obtained by reading the `__rtti_base` global. Essentially, the compiler maps every concrete class to a unique id, starting with 0 \(=ArrayBuffer\), 1 \(=String\) and 2 \(=ArrayBufferView\) . For each such class, the compiler remembers the id of the respective base class, if any, and a set of flags describing the class. Flags for example contain information about key and value alignments, whether a class is managed and so on. Structure is like this:
+It is also possible to build custom data types that integrate with the GC by implementing the `__visit` interface (iterates all references contained in an object of this kind). Please refer to the existing data types' sources for examples.
```
romdotdog:
```suggestion
It is also possible to build custom data types that integrate with the GC by implementing the `__visit` interface (that iterates all references contained in an object of this kind). Please refer to the sources of preexisting data types for examples.
```

src/runtime.md
```diff
@@ -4,160 +4,114 @@ description: Because nobody enjoys being bullied by malloc, free and friends.
...
-| Name        | Offset | Type | Description
-| :---------- | -----: | :--- | :----------
-| #count      |      0 | u32  | Number of concrete classes
-| #flags\[0\] |      4 | u32  | Flags describing the class with id=0
-| #base\[0\]  |      8 | u32  | Base class id of the class with id=0
-| #flags\[1\] |     12 | u32  | ... etc. ...
+## Calling convention
 
-Flag values are currently still in flux, but if you are interested in these, feel free to take a look at the sources.
+AssemblyScript's calling convention is relatively straight forward, as it does not add additional parameters to functions or similar. Note that generated bindings take care of the calling convention automatically, but other environments may want to adhere to it specifically.
```
romdotdog:
```suggestion
AssemblyScript's calling convention is relatively straight forward, as it does not add additional parameters to functions or other behind-the-scenes magic. Note that generated bindings take care of the calling convention automatically, but other environments may want to adhere to it specifically.
```

src/runtime.md
```diff
@@ -4,160 +4,114 @@ description: Because nobody enjoys being bullied by malloc, free and friends.
...
 
-#### Notes
+### Basic values
 
-Unlike other reference counting implementions, allocating an object starts with a reference count of 0. This is useful in standard library code where memory is first allocated and then assigned to a variable, establishing the first reference.
+Exported functions wrap basic numeric return values to their respective value ranges. Basic numeric values passed to exported functions are wrapped to their respective value ranges on demand. Functions internal to the module do not provide these guarantees.
```
romdotdog: > Functions internal to the module do not provide these guarantees.

I don't know what this means

src/runtime.md
```diff
@@ -4,160 +4,114 @@ description: Because nobody enjoys being bullied by malloc, free and friends.
...
+### Managed values
 
-Due to the lack of random access to WebAssembly's execution stack \(remember: AssemblyScript does not have a stack on its own\), values returned from a function become pre-retained for the caller, expecting that it will release the reference at the end, or give the reference to its own caller. This also leads to situations where sometimes a reference must be retained on assignment, and sometimes it must not. It also leads to situations where two branches must be unified, for example if there is a pre-retained value in one arm and one that is not in the other. The compiler is taking care of these situations, but it also requires special care when dealing with the reference counting internals within the standard library.
+Any kind of object is passed as a pointer into memory, and the host is expected to perform the steps necessary to exchange the value between linear memory and the host system. Passing a string for example passes a pointer to its first char code, and the managed header can be evaluated to obtain the string's byte length.
```
romdotdog:
```suggestion
Any kind of object is passed as a pointer into memory, and the host is expected to perform the steps necessary to exchange the value between linear memory and the host system. Passing a string, for example, is composed of passing a pointer to its first char code, and using the managed header to obtain the string's byte length.
```

src/runtime.md
```diff
@@ -4,160 +4,114 @@ description: Because nobody enjoys being bullied by malloc, free and friends.
...
+When an exported function allows one or multiple arguments to be omitted, the module must be informed of the number of significant arguments by calling `exports.__setArgumentsLength(numArgs)` before calling the export, so it can fill in default values. Omitted arguments are not evaluated, zeroes can be passed. Not calling the helper leads to undefined behavior. If a function has a fixed number of arguments, it is not necessary to call the helper. The helper may not be present if a module only exports functions with a fixed number of arguments.
```
romdotdog:
```suggestion
When an exported function allows one or multiple arguments to be omitted, the module must be informed of the number of significant arguments by calling `exports.__setArgumentsLength(numArgs)` before calling the export, so it can fill in default values. Omitted arguments are not evaluated, i.e. zeroes can be passed. Not calling the helper leads to undefined behavior. If a function has a fixed number of arguments, it is not necessary to call the helper. The helper may not be present if a module only exports functions with a fixed number of arguments.
```

src/stdlib/globals.md
```diff
@@ -39,3 +39,1303 @@ In addition to [low-level WebAssembly and compiler built-ins](./builtins.md), th
   function parseFloat(str: string): f64
   ```
   Parses a string to a 64-bit float. Returns `NaN` on invalid inputs.
+
+## Builtins
+
+The following builtins provide direct access to WebAssembly and compiler features. They form the low-level foundation of the standard library, for example, while also being available for everyone to utilize where directly tapping into WebAssembly or the compiler is desired.
```
romdotdog:
```suggestion
The following builtins provide direct access to WebAssembly and compiler features. They form the low-level foundation of the standard library, while also being available for everyone to utilize where directly tapping into WebAssembly or the compiler is desired.
```

src/stdlib/globals.md
```diff
@@ -39,3 +39,1303 @@ In addition to [low-level WebAssembly and compiler built-ins](./builtins.md), th
...
+  if (isString<T>()) { // eliminated if T is not a string
+    return parseInt(a) + parseInt(b)
+  } else { // eliminated if T is a string
+    return a + b
+  }
+}
+```
+
+::: tip
+If you are not going to use low-level WebAssembly in the foreseeable future, feel free to come back to the following paragraphs at a later time and continue at the next page[ ](./loader.md)right away.
```
romdotdog:
```suggestion
If you are not going to use low-level WebAssembly in the foreseeable future, feel free to come back to the following paragraphs at a later time and [continue to the next page](./loader.md) right away.
```

src/stdlib/globals.md
```diff
@@ -39,3 +39,1303 @@ In addition to [low-level WebAssembly and compiler built-ins](./builtins.md), th
...
+  function isConstant(expression: auto): bool
+  ```
+  Tests if the specified expression evaluates to a constant value. Compiles to a constant.
+
+* ```ts
+  function isManaged<T>(expression: auto): bool
+  ```
+  Tests if the specified type _or_ expression is of a managed type. Compiles to a constant. Usually only relevant when implementing custom collection-like classes.
+
+#### Example
```
romdotdog:
```suggestion
#### Example of static type checking
```

src/stdlib/globals.md
```diff
@@ -39,3 +39,1303 @@ In addition to [low-level WebAssembly and compiler built-ins](./builtins.md), th
...
+::: tip
+If you are not going to use low-level WebAssembly in the foreseeable future, feel free to come back to the following paragraphs at a later time and continue at the next page[ ](./loader.md)right away.
+:::
+
+### Utility
+
+* ```ts
+  function sizeof<T>(): usize
+  ```
+  Determines the byte size of the respective _basic type_. Means: If `T` is a class type, the size of `usize` is returned. To obtain the size of a class in memory, use `offsetof<T>()` instead. Compiles to a constant.
```
romdotdog:
```suggestion
  Determines the byte size of the respective _basic type_. Means: If `T` is a class type, the size of `usize`, the pointer type, is returned. To obtain the size of a class in memory, use `offsetof<T>()` instead. Compiles to a constant.
```

src/stdlib/globals.md
```diff
@@ -39,3 +39,1303 @@ In addition to [low-level WebAssembly and compiler built-ins](./builtins.md), th
...
+
+* ```ts
+  function offsetof<T>(fieldName?: string): usize
+  ```
+  Determines the offset of the specified field within the given class type. Returns the class type's end offset \(means: where the next field would be located, before alignment\) if field name has been omitted. Compiles to a constant. The `fieldName` argument must be a compile-time constant `string` because there is no information about field names anymore in the final binary. Hence, the field's name must be known at the time the returned constant is computed.
```
romdotdog:
```suggestion
  Determines the offset of the specified field within the given class type. If `fieldName` is omitted, this returns what could be interpreted as either the size of the class, or the offset where the next field would be located, before alignment. Compiles to a constant. The `fieldName` argument must be a compile-time constant `string` because there is no information about field names anymore at runtime. Therefore, the field's name must be known at the time the returned constant is computed.
```

src/stdlib/globals.md
```diff
@@ -39,3 +39,1303 @@ In addition to [low-level WebAssembly and compiler built-ins](./builtins.md), th
...
+
+* ```ts
+  function alignof<T>(): usize
+  ```
+  Determines the alignment \(log2\) of the specified underlying _basic type_. Means: If `T` is a class type, the alignment of `usize` is returned. Compiles to a constant.
```
romdotdog:
```suggestion
  Determines the alignment \(log2\) of the specified underlying _basic type_; i.e. if `T` is a class type, the alignment of `usize` is returned. Compiles to a constant.
```

src/stdlib/globals.md
```diff
@@ -39,3 +39,1303 @@ In addition to [low-level WebAssembly and compiler built-ins](./builtins.md), th
...
+
+* ```ts
+  function assert<T>(isTrueish: T, message?: string): T
+  ```
+  Traps if the specified value is not true-ish, otherwise returns the non-nullable value. Like assertions in C, aborting the entire program if the expectation fails, with the `--noAssert` option to disable all assertions in production.
```
romdotdog:
```suggestion
  Traps if the specified value is not true-ish, otherwise returns the non-nullable value. Like assertions in C, aborting the entire program if the expectation fails. There is a [compiler `--noAssert` option](./compiler.md) to disable all assertions in production.
```

src/stdlib/globals.md
```diff
@@ -39,3 +39,1303 @@ In addition to [low-level WebAssembly and compiler built-ins](./builtins.md), th
...
+
+* ```ts
+  function idof<T>(): u32
+  ```
+  Obtains the computed unique id of a class type. Usually only relevant when allocating objects or dealing with RTTI externally.
+
+* ```ts
+  function nameof<T>(value?: T): string
+  ```
+  Determines the name of a given type.
```
romdotdog:
```suggestion
  Returns the name of a given type.
```

src/stdlib/globals.md
```diff
@@ -39,3 +39,1303 @@ In addition to [low-level WebAssembly and compiler built-ins](./builtins.md), th
...
+    return a + b
+  }
+}
+```
+
+::: tip
+If you are not going to use low-level WebAssembly in the foreseeable future, feel free to come back to the following paragraphs at a later time and continue at the next page[ ](./loader.md)right away.
+:::
+
+### Utility
```
romdotdog: Seems like `Utilities` could be a better name

src/stdlib/globals.md
```diff
@@ -39,3 +39,1303 @@ In addition to [low-level WebAssembly and compiler built-ins](./builtins.md), th
...
+
+  | T                                          | Instruction
+  |--------------------------------------------|-------------
+  | f32                                        | f32.nearest
+  | f64                                        | f64.nearest
+  | i8, u8, i16, u16, i32, u32, i64, u64, bool | *none*
+  </details>
+
+* ```ts
+  function reinterpret<TTo>(value: auto): T
```
romdotdog: Is `TTo` here intended?

src/stdlib/globals.md
```diff
@@ -39,3 +39,1303 @@ In addition to [low-level WebAssembly and compiler built-ins](./builtins.md), th
...
+#### Memory
+
+Similarly, the following built-ins emit WebAssembly instructions accessing or otherwise modifying memory.
+
+* ```ts
+  function load<T>(ptr: usize, immOffset?: usize): T
+  ```
+  <details><summary>Loads a value of the specified type from memory. Equivalent to dereferencing a pointer in other languages.</summary>
+
+  | T        | Instruction  | If context is i64
```
romdotdog: Elaborate what an `i64` context is here?

src/stdlib/globals.md
```diff
@@ -39,3 +39,1303 @@ In addition to [low-level WebAssembly and compiler built-ins](./builtins.md), th
...
+
+* ```ts
+  function memory.repeat(dst: usize, src: usize, srcLength: usize, count: usize): void
+  ```
+  Repeats a sequence of bytes given as `src` with `srcLength` `count` times into destination `dst`.
+
+* ```ts
+  function memory.compare(lhs: usize, rhs: usize, n: usize): i32
+  ```
+  Compares the first `n` bytes of `left` and `rigth` and returns a value that indicates their relationship:
```
romdotdog:
```suggestion
  Compares the first `n` bytes of `left` and `right` and returns a value that indicates their relationship:
```

src/stdlib/globals.md
```diff
@@ -39,3 +39,1303 @@ In addition to [low-level WebAssembly and compiler built-ins](./builtins.md), th
...
+  Gets a pointer to a pre-initialized static chunk of memory. Alignment defaults to the size of `T`. Arguments must be compile-time constants.
+
+The `immOffset` argument is a bit special here, because it becomes an actual immediate of the respective WebAssembly instruction instead of a normal operand. Thus it must be provided as a compile time constant value. This can be a literal or the value of a `const` variable that the compiler can precompute.
+
+#### Control flow
+
+* ```ts
+  function select<T>(ifTrue: T, ifFalse: T, condition: bool): T
+  ```
+  Selects one of two pre-evaluated values depending on the condition. Differs from an `if/else` in that both arms are always executed and the final value is picked based on the condition afterwards. Performs better than an `if/else` only if the condition is random \(means: branch prediction is not going to perform well\) and both alternatives are cheap. It is also worth to note that Binaryen will do relevant optimizations like switching to a `select` automatically, so using a ternary `? :` for example is just fine.
```
romdotdog:
```suggestion
  Selects one of two pre-evaluated values depending on the condition. Differs from an `if/else` in that both arms are always executed and the final value is picked based on the condition afterwards. Performs better than an `if/else` only if the condition is random \(means: branch prediction is not going to perform well\) and both alternatives are cheap. It is also worth to note that Binaryen will do relevant optimizations like switching to a `select` automatically, so using a ternary `? :` may be preferable.
```

src/stdlib/globals.md
```diff
@@ -39,3 +39,1303 @@ In addition to [low-level WebAssembly and compiler built-ins](./builtins.md), th
...
+  function memory.data(size: i32, align?: i32): usize
+  ```
+  Gets a pointer to a zeroed static chunk of memory of the given size. Alignment defaults to `16`. Arguments must be compile-time constants.
+
+* ```ts
+  function memory.data<T>(values: T[], align?: i32): usize
+  ```
+  Gets a pointer to a pre-initialized static chunk of memory. Alignment defaults to the size of `T`. Arguments must be compile-time constants.
+
+The `immOffset` argument is a bit special here, because it becomes an actual immediate of the respective WebAssembly instruction instead of a normal operand. Thus it must be provided as a compile time constant value. This can be a literal or the value of a `const` variable that the compiler can precompute.
```
romdotdog: Surround this in a tip or warning?

src/stdlib/globals.md
```diff
@@ -39,3 +39,1303 @@ In addition to [low-level WebAssembly and compiler built-ins](./builtins.md), th
...
+* ```ts
+  function atomic.cmpxchg<T>(ptr: usize, expected: T, replacement: T, immOffset?: usize): T
+  ```
+  Atomically compares and exchanges an integer value in memory if the condition is met.
+
+* ```ts
+  function atomic.wait<T>(ptr: usize, expected: T, timeout: i64): AtomicWaitResult
+  ```
+
+  Performs a wait operation on an address in memory suspending this agent if the integer condition is met. Return values are
```
romdotdog:
```suggestion
  Performs a wait operation on an address in memory, suspending this agent if the integer condition is met. Return values are
```

src/stdlib/globals.md
```diff
@@ -39,3 +39,1303 @@ In addition to [low-level WebAssembly and compiler built-ins](./builtins.md), th
...
+  function atomic.notify(ptr: usize, count: i32): i32
+  ```
+  Performs a notify operation on an address in memory waking up suspended agents.
+
+* ```ts
+  function atomic.fence(): void
+  ```
+  Performs a fence operation, preserving synchronization guarantees of higher level languages.
+
+Again, the `immOffset` argument must be a compile time constant value.
```
romdotdog: Surround this with a tip here too or maybe even remove?

src/stdlib/globals.md
```diff
@@ -39,3 +39,1303 @@ In addition to [low-level WebAssembly and compiler built-ins](./builtins.md), th
...
+  | i32 | i32x4.max_s
+  | u32 | i32x4.max_u
+  | f32 | f32x4.max
+  | f64 | f64x2.max
+  </details>
+
+* ```ts
+  function v128.pmin<T>(a: v128, b: v128): v128
+  ```
+  <details><summary>Computes the psuedo-minimum of each lane.</summary>
```
romdotdog, line 923: Elaborate on what these "pseudo" minimums and maximums are?

src/stdlib/globals.md
```diff
@@ -39,3 +39,1303 @@ In addition to [low-level WebAssembly and compiler built-ins](./builtins.md), th
...
+  | T   | Instruction
+  |-----|-------------
+  | f32 | f32x4.trunc
+  | f64 | f64x2.trunc
+  </details>
+
+* ```ts
+  function v128.nearest<T>(a: v128): v128
+  ```
+  <details><summary>Rounds to the nearest integer tied to even of each lane.</summary>
```
romdotdog: What does this mean?

src/stdlib/string.md
```diff
@@ -222,4 +222,14 @@ The following mostly exist to have a safe way to copy between Strings and ArrayB
 
 ## Considerations
```
romdotdog, line 224: Provide a TL;DR here before throwing docs readers a huge block of text.

romdotdog (changes requested):
Nice job. Just a few (50) changes.

src/introduction.md
dcodeIO: Right, it's a pun on JavaScript obviously not being Java and AssemblyScript obviously not being Assembly :)

src/stdlib/globals.md
dcodeIO: Yes, denotes that the output type (the type reinterpreted "to") is provided, not the input type. Missed that the returned `T` should be `TTo` as well.

src/stdlib/globals.md
dcodeIO: This one is taken right from the specification I believe because I couldn't find a proper explanation. Agree that we should explain it in more detail.

src/compiler.md
dcodeIO: Use of the option is relatively rare, so I put it below. Not sure.

src/introduction.md
```diff
@@ -46,22 +40,53 @@ For comparison, the following C code is roughly equivalent:
...
+Both perspectives can be mixed depending on whether low-level control with WebAssembly instructions or idiomatic concepts with the managed standard library are desirable to accomplish an individual task.
+
+## Frequently asked questions
+
+::: tip Does AssemblyScript involve an interpreter, or a "VM in a VM"?
+No, AssemblyScript compiles to WebAssembly bytecode directly, statically, ahead-of-time.
+:::
+
+::: tip What are the differences between AssemblyScript and TypeScript?
+TypeScript transpiles down to JavaScript, a dynamic just-in-time compiled language. AssemblyScript, on the other hand, compiles to a static WebAssembly binary. Their implementation is wildly, and their feature set is slightly different. However, the two languages are so very similar on the surface that they share many concepts. For example, TypeScript tooling can be used to author and refactor AssemblyScript code and, with some effort, the same code base can be transpiled to JavaScript with `tsc` and compiled to WebAssembly with `asc`, or code shared. The AssemblyScript compiler itself is portable.
```
romdotdog: > Their implementation is wildly, and their feature set is slightly different.

wildly what?

src/compiler.md
```diff
@@ -191,27 +153,348 @@ Other options include those forwarded to Binaryen and various flags useful in ce
...
+
+### Portable standard library
+
+Besides the full standard library, AssemblyScript provides a portable variant of the functionality that is present in both JavaScript and WebAssembly. In addition to that, the portable library lifts some of the functionality that is only available with `asc` to JavaScript, like the portable conversions mentioned below.
+
+Also note that some parts of JavaScript's standard library function a little more loosely than how they would when compiling to WebAssembly. While the portable definitions try to take care of this, one example where this can happen is `Map#get` returning `undefined` when a key cannot be found in JavaScript, while resulting in an abort in WebAssembly, where it is necessary to first check that the key exists using `Map#has`.
+
+To use the portable library, add the following somewhere along your build step so the portable features are present in the environment
```
romdotdog:
```suggestion
To use the portable library, add the following somewhere along your build step so the portable features are present in the environment:
```

romdotdog (changes requested):

src/runtime.md
```diff
@@ -4,160 +4,122 @@ description: Because nobody enjoys being bullied by malloc, free and friends.
...
+| mmInfo   |    -20 | usize | Memory manager info
+| gcInfo   |    -16 | usize | Garbage collector info
+| gcInfo2  |    -12 | usize | Garbage collector info
+| rtId     |     -8 | u32   | Unique id of the concrete class
+| rtSize   |     -4 | u32   | Size of the data following the header
+|          |        |       |
+|          |      0 |       | Payload starts here
 
-Both the TLSF memory manager and the concept of reference counting are a good fit for _predictable performance_ scenarios. Remember: WebAssembly is about exactly that. It is not necessarily the fastest choice in every possible scenario, though.
+References to an object always point at the start of the payload, with the header beginning in the 20 bytes before. Null references are just the value `0`. When working with an AssemblyScript module externally, knowing the memory layout can be helpful to for example obtain an object's class id or size. Invoking `__new` automatically prepends a managed header and registers the object with the GC, using the provided class id for `rtId` and the provided size for `rtSize`.
```
romdotdog:
```suggestion
References to an object always point at the start of the payload, with the header beginning 20 bytes before. Null references are just the value `0`. When working with an AssemblyScript module externally, knowing the memory layout can be helpful to for example obtain an object's class id or size. Invoking `__new` automatically prepends a managed header and registers the object with the GC, using the provided class id for `rtId` and the provided size for `rtSize`.
```

src/introduction.md
dcodeIO: > Their implementation is wildly [different], and their feature set is slightly different.

Is it uncommon to omit repetition like this in EN?

PR #141, opened 2022-06-05
Update getting-started.md

furetosan:
Simple grammar

Don't mean to bother


romdotdog:
This is incorrect grammar, afaik. Already, by dictionary, means "so soon" among other things and that's its connotation here.

PR #142, opened 2022-06-05
Mention the `entries` option of `asconfig.json`

trusktr:


dcodeIO:
Looking at this one again, it seems `entries` is already [mentioned here](https://www.assemblyscript.org/compiler.html#configuration-file).

PR #143, opened 2022-06-19
Update built-with-assemblyscript.md

trusktr:
Added some project to built-with, and also organized categories a bit.


trusktr:
I think we should list more projects here, including promising but not-necessarily-finished projects, so that it can serve as an index to where people may want to get involved.

dcodeIO (approved):

PR #176, opened 2022-08-30
Update globals.md

xhyumiracle:
fix "Example of static type checking" comments.

### DIFF
--- a/src/stdlib/globals.md
+++ b/src/stdlib/globals.md
@@ -112,13 +112,13 @@ By making use of the following special type checks, especially in generic contex
 
 ```ts
 function add<T>(a: T, b: T): T {
-  return a + b // addition if numeric, string concatenation if a string
+  return a + b // addition if numeric, string concatenation if string
 }
 
 function add<T>(a: T, b: T): T {
-  if (isString<T>()) { // eliminated if T is not a string
+  if (isString<T>()) { // eliminated if T is string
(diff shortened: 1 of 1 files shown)


src/stdlib/globals.md
```diff
@@ -112,13 +112,13 @@ By making use of the following special type checks, especially in generic contex
 
 ```ts
 function add<T>(a: T, b: T): T {
-  return a + b // addition if numeric, string concatenation if a string
+  return a + b // addition if numeric, string concatenation if string
 }
 
 function add<T>(a: T, b: T): T {
-  if (isString<T>()) { // eliminated if T is not a string
+  if (isString<T>()) { // eliminated if T is string
```
dcodeIO, line 119: Might be a little unclear, but the comments here refer to the respective branch, which in this case is eliminated if `T` is **not** a string. Then the `else` branch will be the only one compiled.

src/stdlib/globals.md
```diff
@@ -112,13 +112,13 @@ By making use of the following special type checks, especially in generic contex
...
     return parseInt(a) + parseInt(b)
-  } else { // eliminated if T is a string
+  } else { // eliminated if T is not string
```
dcodeIO, line 121: Similarly, the `else` branch is eliminated if `T` **is** a string, whereas the prior branch will be compiled.

PR #182, opened 2022-10-20
Initial docs on reference / gc / string types

dcodeIO:
Adds some initial documentation on future Wasm GC and related types. Mostly useful for experimentation currently.


MaxGraey (approved):
nice!

PR #187, opened 2022-12-20
correct chicken and the egg idiom

ccouzens:
https://idioms.thefreedictionary.com/chicken+and+egg

I'm British and I've never heard it expressed henn and egg situation before.

### DIFF
--- a/src/concepts.md
+++ b/src/concepts.md
@@ -166,7 +166,7 @@ The respective implementations of `abort`, `trace` and `seed` can be overridden 
 
 ### Accessing memory during instantiation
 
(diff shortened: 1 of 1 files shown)


CountBleck:
You're absolutely right, but did we need to know you're British? 😜

dcodeIO:
Origin: In German, it's [Henne-Ei-Problem](https://de.wikipedia.org/wiki/Henne-Ei-Problem), and the original author (me) messed up. Twice, because "hen".

dcodeIO (approved):

dcodeIO:
Thanks :)

PR #205, opened 2023-11-29
fix: wrong module path

oliyg:
Compiler `asc.js` for node.js is under the path `assemblyscript/cli`. So we should modify this content by changing `assemblyscript/asc` to `assemblyscript/cli/asc`.

### DIFF
--- a/src/compiler.md
+++ b/src/compiler.md
@@ -231,7 +231,7 @@ asc --config asconfig.json --target release
 The compiler API can also be used programmatically:
 
 ```js
-import asc from "assemblyscript/asc";
+import asc from "assemblyscript/cli/asc";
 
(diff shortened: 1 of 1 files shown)


dcodeIO:
The assemblyscript package has a dedicated `asc` export to pick up the built compiler, and that's used here:

https://github.com/AssemblyScript/assemblyscript/blob/1e0466ef94fa5cacd0984e4f31a0087de51538a8/package.json#L47

oliyg:
> The assemblyscript package has a dedicated `asc` export to pick up the built compiler, and that's used here:
> 
> https://github.com/AssemblyScript/assemblyscript/blob/1e0466ef94fa5cacd0984e4f31a0087de51538a8/package.json#L47

You are right.

PR #220, opened 2024-02-24
fix documentation for v128.andnot

Yaffle:


CountBleck (approved):
Looks like that matches the [spec](https://webassembly.github.io/spec/core/exec/numerics.html#xref-exec-numerics-op-iandnot-mathrm-iandnot-n-i-1-i-2).

PR #239, opened 2026-01-29
Fix links for Firefox, Wasmer, and Wasmtime

scottbedard:
These links are stale, I've updated them to the URL I thought looked closest


HerrCai0907 (approved):
LGTM

PR #241, opened 2026-04-11
Update dependencies & simplify monaco cross-domain worker setup

MaxGraey:


src/.vuepress/public/editor.html
```diff
@@ -311,56 +311,31 @@ <h5>Documentation</h5>
...
-          }
-          return window.URL.createObjectURL(blob)
+    function crossDomainWorker() {
+      window.MonacoEnvironment = {
+        getWorkerUrl() {
+          const proxy = `
+            self.MonacoEnvironment = { baseUrl: '${MONACO_BASE_URL}' };
+            importScripts('${MONACO_BASE_URL}vs/base/worker/workerMain.js');
+          `
+          return URL.createObjectURL(new Blob([proxy], { type: 'application/javascript' }))
```
MaxGraey, line 331: @dcodeIO WDYT for such simplification?

dcodeIO: Wondering if some of these things served a real purpose. Say when Monaco runs on some hot reload capable dev server, do we perhaps want to preserve an existing Monaco environment instead of unconditionally creating a new one and leak the old? Or do we need same origin handling in some odd configurations (like in dev)? The previous try-catch there looks like some compatibility fallback, can that still be relevant nowadays?

Iirc, originally I basically copied whatever Monaco documented. Well possible that there is generally a better way today, one that doesn't even use AMD anymore.

MaxGraey: New implementation based on this tutorial:
https://log.schemescape.com/posts/web-development/embedding-monaco-from-cdn.html
and it looks like this typical approach after `monaco-editor` >= `v0.52.x`
https://github.com/egoist/vue-monaco/blob/master/README.md?plain=1#L122

With one exception, I decided to preserve data url creation via `URL.createObjectURL`

dcodeIO: Ah. I was also wondering if there is a more modern ESM-aware way of doing the setup nowadays. Have you checked?

MaxGraey: Only this: https://github.com/microsoft/monaco-editor/tree/main/samples/browser-esm-esbuild

PR #257, opened 2026-06-10
Add Web Application Frameworks section

zion-sati:
Added a section for Web Application Frameworks with fui-as.

@JairusSW Let me know if you want me to put it under an existing section - just couldn't work out which one is the best for fui-as which is a Web App framework. 🙏


MaxGraey:
Thanks!

